PR #12, opened 2019-05-14
[WIP] Add the pin_project attribute

taiki-e:
Add a new attribute macro. This works as follows:

```rust
#[pin_project(self)]
fn foo(mut self: Pin<&mut Self>) {
    let x: Pin<&mut i32> = self.field1;
    let y: &mut i32 = self.field2;
}

// desugaring
fn foo(self: Pin<&mut Self>) {
    let mut __self = self.project();
    let x: Pin<&mut i32> = __self.field1;
    let y: &mut i32 = __self.field2;
}
```

TODO: write docs

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -147,6 +147,8 @@ extern crate proc_macro;
 mod utils;
 
 #[cfg(feature = "project_attr")]
+mod pin_project;
+#[cfg(feature = "project_attr")]
 mod project;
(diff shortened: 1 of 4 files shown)


taiki-e:
I don't plan to put this feature in the next version. [In the next version](https://github.com/taiki-e/pin-project/pull/21), we will stick to making pin-project safer.

PR #20, opened 2019-08-04
Set up CI with Azure Pipelines

taiki-e:


taiki-e:
This should not conflict with #18.

Issue #36, opened 2019-08-10
Should #[project] attribute be disabled by default?


taiki-e:
This is an attribute to support enum pin-projections (https://github.com/rust-lang-nursery/pin-utils/issues/21, [docs](https://docs.rs/pin-project/0.3.4/pin_project/attr.unsafe_project.html#enums)) and destructing of pin-projected structs, but since it depends on the "syn/visit-mut" feature, there is a possibility of increasing the compilation time.
It may be preferable to be disabled by default if it is rarely used.

taiki-e:
cc @Nemo157

PR #37, opened 2019-08-10
Remove support of if-let expressions from #[project] attribute

taiki-e:
This has many limitations.

cc #36


taiki-e:
bors retry

PR #42, opened 2019-08-11
Release 0.4.0-alpha.1

taiki-e:
[Changelog](https://github.com/taiki-e/pin-project/blob/1276ad933cf5f0b2aad98a0d38fd4cbe0634090e/CHANGELOG.md#040-alpha1---2019-08-11)

cc #21


src/lib.rs
```diff
@@ -1,4 +1,10 @@
-//! An attribute that creates a projection struct covering all the fields.
+//! A crate for safe and ergonomic pin-projection.
+//!
+//! This crate provides the following attribute macros:
+//!
+//! * [`pin_project`] - An attribute that creates a projection struct covering all the fields.
```
taiki-e: This is not a better way, but I couldn't come up with any other good way.

pin-project-internal/src/lib.rs
```diff
@@ -18,20 +20,373 @@ mod pinned_drop;
 mod project;
 
 use proc_macro::TokenStream;
-
 use utils::Nothing;
 
+// TODO: Move this doc into pin-project crate when https://github.com/rust-lang/rust/pull/62855 merged.
```
taiki-e: https://github.com/taiki-e/pin-project/pull/42/files#r312743458

taiki-e:
cc @Aaron1011

PR #46, opened 2019-08-18
Add impl block support to #[project] attribute

### LINKED ISSUE #43: Provide way to refer to projected type
A couple of things I've wanted to do is to be able to destructure the projection to avoid using `this.` all the time in methods, and implement helper methods directly on the projection. As an example this currently works because of a lack of hygiene:

```rust
#[pin_project::pin_project]
struct Foo {
    #[pin]
    bar: &'static str,
}

impl __FooProjection<'_> {
    fn debug(&self) {
        println!("{:?}", self.bar);
    }
}

fn main() {
    let mut foo = Box::pin(Foo { bar: "baz" });
    foo.as_mut().project().debug();
    let __FooProjection { bar } = foo.as_mut().project();
    println!("{}", bar);
}
```

but it would be nice to have a way to do this that doesn't rely on the internal details of the macro.

taiki-e:
This adds a feature to convert an annotated impl to a projected type's impl.

Example:

```rust
#[project]
impl<T, U> Foo<T, U> {
    fn baz(self) {
        let Self { future, field } = self;
    }
}

// convert to:
impl<'_pin, T, U> __FooProjection<'_pin, T, U> {
    fn baz(self) {
        let Self { future, field } = self;
    }
}
```

Closes #43

cc @Nemo157

### DIFF
--- a/pin-project-internal/src/lib.rs
+++ b/pin-project-internal/src/lib.rs
@@ -289,7 +289,7 @@ pub fn pin_project(args: TokenStream, input: TokenStream) -> TokenStream {
 /// `Pin<&mut Self>`. In particular, it will never be called more than once,
 /// just like [`Drop::drop`].
 ///
-/// Example:
+/// ## Example
 ///
 /// ```rust
 /// use pin_project::{pin_project, pinned_drop};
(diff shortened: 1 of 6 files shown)


Nemo157:
I only looked at the docs and tests, but design looks good to me!

PR #56, opened 2019-09-01
Release 0.4.0-alpha.6

taiki-e:

PR #62, opened 2019-09-03
Improve document of UnpinStruct

taiki-e:
Improve document and add a way to opt-out mentioned in https://github.com/taiki-e/pin-project/pull/53#issuecomment-525974922.

The reason for needing support for ways other than `-Zallow-features` is that it also affects the use of other dependency unstable features.

Related:  #54 https://github.com/taiki-e/pin-project/pull/56#discussion_r319740800

Generated document:

struct-1
struct-2

Ways to opt-out of this:

https://github.com/taiki-e/pin-project/blob/5f7c6e35c2a9f3dfbc01a63e6206875d1bfb986b/pin-project-internal/build.rs#L21-L28


taiki-e:
cc @Aaron1011 @Nemo157 @LucioFranco

taiki-e:
bors r=Aaron1011

PR #70, opened 2019-09-04
Remove `_` from UnpinStruct

taiki-e:
[As `UnpinStruct` is displayed in the documents](https://github.com/taiki-e/pin-project/pull/62), I think it looks better when `_` is removed.

Before (https://docs.rs/hyper/0.13.0-alpha.1/hyper/server/index.html):
docs-1

After:
docs-2

cc @seanmonstar @LucioFranco

### DIFF
--- a/pin-project-internal/src/pin_project/mod.rs
+++ b/pin-project-internal/src/pin_project/mod.rs
@@ -153,7 +153,7 @@ impl Context {
                 }
             };
 
-            let struct_ident = format_ident!("__UnpinStruct{}", orig_ident, span = make_span());
+            let struct_ident = format_ident!("UnpinStruct{}", orig_ident, span = make_span());
             let always_unpin_ident = format_ident!("AlwaysUnpin{}", orig_ident, span = make_span());
 
             // Generate a field in our new struct for every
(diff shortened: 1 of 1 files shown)


taiki-e:
@LucioFranco
> 👍 wonder if we could provide the option to name it to a custom name.

Looks like a good idea.

Probably like this? 
```rust
#[pin_project(unpin_docs = MyUnpinStructName)]
struct Foo<T, U> {
    #[pin]
    pinned: T,
    unpinned: U,
}
```

(Note: Once rust-lang/rust#48054 is implemented and stabilized, these documents will not be generated, but the options themselves will be maintained for compatibility.)

LucioFranco:
@taiki-e seems reasonable!

seanmonstar:
I didn't even realize that it exports a new type. Could it be `doc(hidden)`, or even better, not `pub`?

taiki-e:
@seanmonstar 
I'm sorry. There is certainly a lack of documentation on this behavior.
This is intended to avoid [the issue that cannot know the actual trait implementation bounds of the `Unpin` implementation from the document of generated code.](https://github.com/rust-lang/rust/issues/63281) And a way to opt-out of this was added in #62.

taiki-e:
So, I am considering disabling this by default.
This is because the current way is that if the struct is declared in a public module, UnpinStruct is displayed, but if it is declared in a private module, there is a discrepancy that UnpinStruct does not appear in the document.

cc @Aaron1011

taiki-e:
It should also be a problem that items that are not declared by the user appear in the document without permission.

taiki-e:
Closing this in favor of #71.

PR #71, opened 2019-09-04
Do not display UnpinStruct in the document by default

taiki-e:
There are some problems as mentioned in #70 

This causes the problem that by default the actual Unpin bounds cannot be known from the document if the original struct/enum itself is public (see https://github.com/taiki-e/pin-project/pull/53#issuecomment-525906867 and https://github.com/rust-lang/rust/issues/63281 for more).

This can be enabled using `--cfg pin_project_show_unpin_struct` in RUSTFLAGS.

```toml
# in Cargo.toml
[package.metadata.docs.rs]
rustdoc-args = ["--cfg", "pin_project_show_unpin_struct"]
```

cc @Aaron1011 @seanmonstar @LucioFranco

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -49,7 +49,7 @@ jobs:
         toolchain: nightly
     - script: |
         cargo clean
-        RUSTFLAGS='--cfg compiletest' cargo test -p pin-project --all-features --test compiletest
+        RUSTFLAGS='--cfg compiletest --cfg pin_project_show_unpin_struct' cargo test -p pin-project --all-features --test compiletest
       displayName: compiletest
 
 - job: clippy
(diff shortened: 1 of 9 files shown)


azure-pipelines.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         toolchain: nightly
     - script: |
         cargo clean
-        RUSTFLAGS='--cfg compiletest' cargo test -p pin-project --all-features --test compiletest
+        RUSTFLAGS='--cfg compiletest --cfg pin_project_show_unpin_struct' cargo test -p pin-project --all-features --test compiletest
```
taiki-e: cc #35 It becomes even more complicated :(

PR #77, opened 2019-09-07
Move automatic generation of Unpin implementation to proc-macro-derive

### LINKED ISSUE #68: pin-project does not interoperate well with #[cfg()]
Take this struct for instance:
```rust
#[pin_project]
#[derive(Debug)]
pub struct Socket {
    #[cfg_attr(unix, pin)]
    #[cfg(unix)]
    inner: tokio::net::unix::UnixStream,
    #[cfg_attr(windows, pin)]
    #[cfg(windows)]
    inner: tokio_named_pipes::NamedPipe,
}
```
I get the error
```rust
error[E0124]: field `inner` is already declared
  --> src/io/socket.rs:29:5
   |
26 |     inner: tokio::net::unix::UnixStream,
   |     ----------------------------------- `inner` first declared here
...
29 |     inner: tokio_named_pipes::NamedPipe,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ field already declared
```
Even though, as far as I can tell, this _should_ work. My guess here is that `pin-project` is run before the `#[cfg()]`s are evaluated, and ends up not propagating them, but I'm not sure?

taiki-e:
To generate the correct `Unpin` implementation, we need to collect the
types of the pinned fields. However, since proc-macro-attribute is
applied before cfg, we cannot be collecting field types at this timing.
So instead of generating the `Unpin` implementation here, we need to
delegate automatic generation of the `Unpin` implementation to
proc-macro-derive.

Fixes #68


compiletest.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# A script to run compile tests with the same condition that pin-project executes with CI.
```
taiki-e: cc #35

pin-project-internal/src/pin_project/enums.rs
```diff
@@ -86,18 +91,27 @@ fn named(
...
-        if let Some(attr) = attrs.find_remove(PIN) {
-            let _: Nothing = syn::parse2(attr.tokens)?;
-            cx.push_unpin_bounds(ty.clone());
+        let cfg = collect_cfg(attrs);
+        if cx.find_pin_attr(attrs)? {
             let lifetime = &cx.lifetime;
-            proj_fields.push(quote!(#ident: ::core::pin::Pin<&#lifetime mut #ty>));
-            proj_body.push(quote!(#ident: ::core::pin::Pin::new_unchecked(#ident)));
+            proj_fields.push(quote! {
+                #(#cfg)* #ident: ::core::pin::Pin<&#lifetime mut #ty>
```
taiki-e: (In the future, it may be preferable to move these processes to proc-macro-derive.)

taiki-e:
Hmm... [tuple structs don't seem to work well either.](https://dev.azure.com/taiki-e/taiki-e/_build/results?buildId=767&view=logs&jobId=21f679a1-fe27-585c-888f-0e663b32681b&j=6ea23a02-24c8-5941-a955-2c76066a2c02)

tests/ui/cfg/field_sneaky.rs
```diff
@@ -0,0 +1,19 @@
...
+#![feature(optin_builtin_traits)]
+#![feature(trivial_bounds)]
+
+#[macro_use]
+extern crate sneaky_macro;
+
+use pin_project::pin_project;
+
+#[pin_project] //~ ERROR pattern does not mention field `__field`
+#[add_pinned_field]
```
taiki-e: I have never seen a crate actually attacked like this, but it should be preferable to be able to detect it.

PR #85, opened 2019-09-10
Add use statements support to #[project] attribute

### LINKED ISSUE #80: Provide way to import projection trait
The current project/project_into methods are trait-methods, so it needs to use glob import to use it from other modules.

We could probably provide the following way:

```rust
#[pin_project]
struct Foo<T> {
    #[pin]
    field: T,
}
mod bar {
    #[project]
    use super::Foo;
    // convert to
    use super::__FooProjectionTrait;
}
```

taiki-e:
This adds a feature to convert annotated imports to a projected type's imports.

~~For now, this works as an import for both projected type and projection trait.~~ Now, this works as an import for projected type.

### Examples

```rust
use pin_project::pin_project;

#[pin_project]
struct Foo<A> {
    #[pin]
    field: A,
}

mod bar {
    use pin_project::project;
    use super::Foo;
    use std::pin::Pin;

    #[project]
    use super::Foo;

    #[project]
    fn baz<A>(foo: Pin<&mut Foo<A>>) {
        #[project]
        let Foo { field } = foo.project();
        let _: Pin<&mut A> = field;
    }
}
```

### DIFF
--- a/pin-project-internal/src/project.rs
+++ b/pin-project-internal/src/project.rs
@@ -23,6 +23,7 @@ fn parse(mut stmt: Stmt) -> Result<TokenStream> {
         Stmt::Local(local) => Context::default().replace_local(local)?,
         Stmt::Item(Item::Fn(ItemFn { block, .. })) => Dummy.visit_block_mut(block),
         Stmt::Item(Item::Impl(item)) => replace_item_impl(item),
+        Stmt::Item(Item::Use(item)) => replace_item_use(item)?,
         _ => {}
     }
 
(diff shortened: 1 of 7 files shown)


taiki-e:
I feel that if importing both the projected-type and the original type, it will not look very good.

```rust
#[project]
use super::Foo;
use super::Foo;
```

taiki-e:
Blocked on https://github.com/taiki-e/pin-project/issues/21#issuecomment-530206201

pin-project-internal/src/project.rs
```diff
@@ -197,3 +204,32 @@ impl VisitMut for Dummy {
...
+
+        match node {
+            // Desugar `use tree::<name>` into `tree::__<name>Projection`.
+            UseTree::Name(name) => replace_ident(&mut name.ident),
+            UseTree::Glob(glob) => {
+                self.res =
+                    Err(error!(glob, "#[project] attribute may not be used on glob imports"));
+            }
+            UseTree::Rename(rename) => {
+                // TODO: Consider allowing the projected type to be renamed by `#[project] use Foo as Bar`.
```
taiki-e: cc #43

taiki-e:
bors retry

PR #90, opened 2019-09-11
Change the argument type of project method back to `self: Pin<&mut Self>`

### LINKED ISSUE #89: Change the argument type of project method back to `self: Pin<&mut Self>`
The signature of the generated `project` method changes as follows.

before:
```rust
fn project(self: &mut Pin<&mut Self>) -> ProjectedType;
```

after:
```rust
fn project(self: Pin<&mut Self>) -> ProjectedType;
```

---

The advantage of the current approach is that you can call the `project` method multiple times without `.as_mut()`.

However, there are drawbacks such as requiring a different method to avoid lifetime issues (#65), and not being able to call the method without importing trait (https://github.com/taiki-e/pin-project/issues/21#issuecomment-530206201).

Also, by reverting this, the code generated by the macro will be simpler.

Related: https://github.com/taiki-e/pin-project/pull/47#issuecomment-530467225

PR: #90

taiki-e:
Closes #89

### DIFF
--- a/tests/pin_project.rs
+++ b/tests/pin_project.rs
@@ -20,7 +20,7 @@ fn test_pin_project() {
     let mut foo = Foo { field1: 1, field2: 2 };
 
     let mut foo_orig = Pin::new(&mut foo);
-    let foo = foo_orig.project();
+    let foo = foo_orig.as_mut().project();
 
     let x: Pin<&mut i32> = foo.field1;
     assert_eq!(*x, 1);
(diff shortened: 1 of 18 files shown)


tests/ui/project/stmt_expr_attributes.rs
```diff
@@ -19,8 +19,7 @@ fn project_stmt_expr_nightly() {
 
     let mut baz = Baz::Variant1(1, 2);
 
-    let mut baz = Pin::new(&mut baz);
-    let mut baz = baz.project();
+    let mut baz = Pin::new(&mut baz).project();
```
taiki-e: Well, this might be useful too.

tests/pin_project.rs
```diff
@@ -295,30 +291,30 @@ fn lifetime_project() {
 
     impl<T, U> Struct<T, U> {
         fn get_pin_mut<'a>(self: Pin<&'a mut Self>) -> Pin<&'a mut T> {
-            self.project_into().pinned
+            self.project().pinned
```
taiki-e: I didn't see this as a problem because I know both ways, but it would be easier to do all of this with a single method.

taiki-e:
bors retry

PR #135, opened 2019-10-13
Move most of the processing to proc-macro-derive

### LINKED ISSUE #123: Make #[pin_project] interoperable with [cfg_attr()]
It would be nice to be able to support the following code:

```rust
#[cfg_attr(feature = "std-future", pin_project)]
struct Foo {
    #[cfg_attr(feature = "std-future", pin)]
    inner: T,
}
```

This is not currently supported for the same reason as https://github.com/taiki-e/pin-project/issues/68#issuecomment-528318870, but I think it can be supported by moving most of the process to proc-macro-derive ~~or parsing `cfg_attr`~~.

Related: https://github.com/tokio-rs/tracing/pull/374#discussion_r331771487

taiki-e:
To generate the correct `Unpin` implementation and the projection methods, we need to collect the types of the pinned fields.  However, since proc-macro-attribute is applied before `#[cfg]` and `#[cfg_attr]` on fields, we cannot be collecting field types properly at this timing. So instead of generating the `Unpin` implementation and the projection methods here, delegate their processing to proc-macro-derive.

#77 moved only the automatic generation of `Unpin` implementation to proc-macro-derive but found that it could not support `#[cfg_attr]`, so this moves more processing.

See also https://github.com/taiki-e/pin-project/issues/68#issuecomment-528318870.

This also adds support for the use of `#[cfg]` on tuple structs and tuple variants.

Closes #123


pin-project-internal/src/pin_project/attribute.rs
```diff
@@ -1,621 +1,71 @@
...
+        #[pin(#private(#args))]
+    });
+
+    Ok(quote! {
+        #(#attrs)*
+        #body
+    })
 }
 
 #[allow(dead_code)] // https://github.com/rust-lang/rust/issues/56750
```
taiki-e: It seems to have been fixed in the latest nightly, but it is left for compatibility with other versions.

PR #158, opened 2019-10-21
Fix compile error with `dyn` types

oxalica:
Add parentheses around original types in generated code.

Currently, it will cause a compile error when field type is like `dyn Trait1 + Trait2`.
```rust
use pin_project::pin_project;
#[pin_project]
struct Struct {
    a: i32,
    #[pin]
    f: dyn std::fmt::Debug + Send,
}
```
will result in,
```
error: ambiguous `+` in a type
 --> src/main.rs:6:8
  |
6 |     f: dyn std::fmt::Debug + Send,
  |        ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use parentheses to disambiguate: `(dyn std::fmt::Debug + Send)`

error: proc-macro derive produced unparseable tokens
 --> src/main.rs:2:1
  |
2 | #[pin_project]
  | ^^^^^^^^^^^^^^
```

### DIFF
--- a/pin-project-internal/src/pin_project/derive.rs
+++ b/pin-project-internal/src/pin_project/derive.rs
@@ -370,10 +370,10 @@ impl<'a> Context<'a> {
 
                 let lifetime = &self.proj.lifetime;
                 proj_fields.push(quote! {
-                    #vis #ident: ::core::pin::Pin<&#lifetime mut #ty>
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks!

bors r+

taiki-e:
bors retry 
(seems bors crashed)

taiki-e:
Published in 0.4.5.

PR #172, opened 2019-12-02
Fix ui test failure

taiki-e:


taiki-e:
perhaps we should pin rustc version...

bors r+

PR #187, opened 2020-04-11
Fix CI failure

taiki-e:


taiki-e:
I can see that [CI has been broken for about 9 days](https://github.com/taiki-e/pin-project/actions/runs/68681662), but it seems GHA didn't send me emails or notifications (but, I received many emails/notifications about cargo-hack's scheduled build failure...). I will look it a bit more...

PR #192, opened 2020-04-18
Hide all generated items except for projected types from calling code

taiki-e:
Generate all items except for projected types inside a `const` scope. This makes it impossible for user code to refer to these types. This removes redundant prefix/postfix based on the original type from many generate types. This also remove the `pin_project_show_unpin_struct` flag.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,12 +72,10 @@ jobs:
       - name: cargo test
         run: |
           cargo test --all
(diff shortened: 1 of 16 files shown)


taiki-e:
> It needs to decide which remove support of rust older than 1.37 and whether to remove the `pin_project_show_unpin_struct` flag.

Keep support for 1.33, remove `pin_project_show_unpin_struct` flag. If someone needed that flag, we can probably restore it later.

PR #195, opened 2020-04-28
Add expansion tests

taiki-e:
Similar to ui tests, but instead of checking the compiler output, this checks the code generated by proc-macro.

Refs: https://github.com/eupn/macrotest

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -22,3 +22,7 @@ cargo +nightly doc --no-deps --all --all-features
 
 echo "Running 'compiletest'"
 . ./compiletest.sh
+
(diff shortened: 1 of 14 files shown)


ci.sh
```diff
@@ -22,3 +22,7 @@ cargo +nightly doc --no-deps --all --all-features
 
 echo "Running 'compiletest'"
 . ./compiletest.sh
+
+# See also https://docs.rs/macrotest/1/macrotest/#updating-expandedrs
```
taiki-e, line 26: Note that the way of updating the generated code is different than trybuild.

PR #219, opened 2020-05-09
Implement !Unpin option

### LINKED ISSUE #108: Provide a way to guarantee the type is !Unpin

taiki-e:
## Example

```rust
#[pin_project(!Unpin)]
struct Struct<T, U> {
    #[pin]
    pinned: T,
    unpinned: U,
}
```

Also, bump MSRV to 1.34 because `unrestricted_attribute_tokens` requires 1.34 (#202 also requires 1.34).

Closes #108


taiki-e:
> It's been more than a year since [1.34 was released](https://blog.rust-lang.org/2019/04/11/Rust-1.34.0.html), and most of the users of this crate depend on futures_api that requires 1.36 (and many of them depend on async_await that requires 1.39). so bumping MSRV to 1.34 is probably no problem.

I think we need to decide on a policy when we release 1.0, but for now, I would like to support a year ago rustc (as long as it is possible).

PR #243, opened 2020-06-06
Allow `project_replace` argument to be used without `Replace`

taiki-e:
Currently, both `Replace` and `project_replace = <name>` arguments are required to name the return value of `project_replace()` method.

```rust
#[pin_project(Replace, project_replace = EnumProjOwn)]
enum Enum<T> {
    Variant(#[pin] T)
}
```

It is not ideal that we need to have both `Replace` and `project_replace = <name>` as `project_replace = <name>` implies that a `project_replace` method is generated.

This PR makes `project_replace` argument to use without `Replace` argument.

```diff
- #[pin_project(Replace, project_replace = EnumProjOwn)]
+ #[pin_project(project_replace = EnumProjOwn)]
  enum Enum<T> {
      Variant(#[pin] T)
  }
```

Also, makes `project_replace` argument an alias for `Replace` so that it can be used without a value.

```rust
#[pin_project(project_replace)]
enum Enum<T> {
    Variant(#[pin] T)
}
```

Related: https://github.com/rust-lang/futures-rs/pull/2176#discussion_r436234178 https://github.com/hyperium/hyper/pull/2220#discussion_r436234137

### DIFF
--- a/examples/project_replace-expanded.rs
+++ b/examples/project_replace-expanded.rs
@@ -5,7 +5,7 @@
 //
 // use pin_project::pin_project;
 //
-// #[pin_project(Replace)]
+// #[pin_project(project_replace)]
 // struct Struct<T, U> {
 //     #[pin]
 //     pinned: T,
--- a/examples/project_replace.rs
+++ b/examples/project_replace.rs
@@ -4,7 +4,7 @@
 
 use pin_project::pin_project;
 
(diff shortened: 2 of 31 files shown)


taiki-e:
Also, the `Replace` argument will be deprecated in the future. (There is no way to generate a warning for an attribute argument, so just remove it in the next major version.)

PR #250, opened 2020-06-08
Preserve the span of self argument

taiki-e:
Refs: https://github.com/dtolnay/async-trait/pull/105

`#[pinned_drop]` attribute also has same issue.


taiki-e:
bor s-

PR #254, opened 2020-06-13
Fix ui tests

taiki-e:


taiki-e:
Running ui tests locally produces these outputs, but never in CI.
I'm not sure why, but I will delete it anyway.

bors r+

PR #261, opened 2020-07-22
Update ui tests to nightly-2020-07-22

taiki-e:


tests/ui/pin_project/project_replace_unsized.stderr
```diff
@@ -6,11 +6,12 @@ error[E0277]: the size for values of type `T` cannot be known at compilation tim
...
   |
-  = help: within `Struct<T>`, the trait `std::marker::Sized` is not implemented for `T`
-  = note: to learn more, visit <https://doc.rust-lang.org/book/ch19-04-advanced-types.html#dynamically-sized-types-and-the-sized-trait>
   = note: required because it appears within the type `Struct<T>`
-  = note: all local variables must have a statically known size
   = help: unsized locals are gated as an unstable feature
+help: function arguments must have a statically known size, borrowed types always have a known size
+  |
+3 | #[pin_project(&project_replace)] //~ ERROR E0277
+  |               ^
```
taiki-e, line 14: 😆

taiki-e: This error message is based on this `Self`:
https://github.com/taiki-e/pin-project/blob/6caecae5adf7ec45f3803aea503d55cac78394bd/pin-project-internal/src/pin_project/derive.rs#L1107

but even if I change the span of this, I get a similar weird message:

```text
help: function arguments must have a statically known size, borrowed types always have a known size
  |
8 | &#[pin_project(project_replace)] //~ ERROR E0277
  | ^
```

The fix requires a different approach.

taiki-e: I've tried some, but for now, I'll leave it as is. (I could remove this help message, but it made other parts a bit worse.)

Issue #262, opened 2020-07-27
?Sized type parameters don't work


alecmocatta:
`?Sized` type parameters don't work unless they are also the last type parameter.

```rust
use pin_project::pin_project;

#[pin_project]
pub struct Foo<'a, I: ?Sized, Item>
where
    I: Iterator,
{
    iter: &'a mut I,
    item: Option<Item>,
}

```

([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=67a8115e207db6514c78ec6fe84081c9))

Errors:

```
   Compiling playground v0.0.1 (/playground)
error[E0277]: the size for values of type `I` cannot be known at compilation time
 --> src/lib.rs:3:1
  |
3 | #[pin_project]
  | ^^^^^^^^^^^^^^ doesn't have a size known at compile-time
4 | pub struct Foo<'a, I: ?Sized, Item>
  |                    - this type parameter needs to be `std::marker::Sized`
  |
  = help: the trait `std::marker::Sized` is not implemented for `I`
  = note: to learn more, visit <https://doc.rust-lang.org/book/ch19-04-advanced-types.html#dynamically-sized-types-and-the-sized-trait>
  = note: only the last element of a tuple may have a dynamically sized type
  = note: this error originates in a derive macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
error: could not compile `playground`.

To learn more, run the command again with --verbose.

```

PR #263, opened 2020-07-27
Fix ?Sized type parameters

alecmocatta:
This is a fix for https://github.com/taiki-e/pin-project/issues/262. Feel free to disregard and implement differently if there is a better fix I'm missing.

Fixes: #262


taiki-e:
Published in 0.4.23

PR #268, opened 2020-09-07
Do not expose unnamed projected types

taiki-e:
Defines projected types in inside `const _: () = {}` when the users don't name it;

cc #264

### DIFF
--- a/examples/enum-default-expanded.rs
+++ b/examples/enum-default-expanded.rs
@@ -35,23 +35,23 @@ where
     Pinned(::pin_project::__private::Pin<&'pin mut (T)>),
     Unpinned(&'pin mut (U)),
 }
-#[doc(hidden)]
-#[allow(dead_code)]
-#[allow(single_use_lifetimes)]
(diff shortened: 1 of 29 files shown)


taiki-e:
This is not actually a breaking change as the generated unnamed projected types aren't public API.

examples/enum-default-expanded.rs
```diff
@@ -35,23 +35,23 @@ where
...
 
 #[doc(hidden)]
 #[allow(non_upper_case_globals)]
 #[allow(single_use_lifetimes)]
 #[allow(clippy::used_underscore_binding)]
 const _: () = {
+    #[allow(dead_code)]
+    #[allow(single_use_lifetimes)]
+    #[allow(clippy::type_repetition_in_bounds)]
+    enum __EnumProjectionRef<'pin, T, U>
```
taiki-e, line 47: This doesn't actually need to be generated, since we can't access the variant without naming it.
I'll optimize it in a later PR.

PR #270, opened 2020-09-08
Optimize code generation when used on enums

taiki-e:
When `#[pin_project]` is used on enums, only named projection types and methods are generated because there is no way to access variants of projected types without naming it. (When `#[pin_project]` is used on structs, both methods are always generated.)

Addresses https://github.com/taiki-e/pin-project/pull/268#discussion_r484313531


taiki-e:
bors retry

PR #273, opened 2020-09-17
Prepare for uplift of drop-bounds lint

### LINKED ISSUE #272: Prepare for uplift of drop-bounds lint
https://github.com/rust-lang/rust/pull/75699 uplifts clippy's `drop_bounds` lint to rustc.
If the implementation is the same as clippy, the code generated by the macro will also be warned (this behavior is probably correct for this lint), and pin-project(-lite) users will be affected by this.

Once https://github.com/rust-lang/rust/pull/75699 or https://github.com/rust-lang/compiler-team/issues/347 merged (or approved), it would be nice to be able to create a release including a patch that allows this lint.

taiki-e:
cc #272

### DIFF
--- a/examples/enum-default-expanded.rs
+++ b/examples/enum-default-expanded.rs
@@ -14,7 +14,7 @@
 // fn main() {}
 // ```
 
-#![allow(dead_code, unused_imports, unused_parens)]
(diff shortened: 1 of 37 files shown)


taiki-e:
Creating pre-release versions frequently is not a very good idea, as pre-release version dependencies are often pinned to that version.
So, it's a bit early, but I'll merge this to include in 1.0.0-alpha.1.

PR #277, opened 2020-09-21
Suppress more lints in generated code

taiki-e:


tests/lint.rs
```diff
@@ -101,13 +112,52 @@ pub mod rustc {
...
+        f: &'pin___ mut T,
+    }
+}
+
+pub mod variant_size_differences {
+    use pin_project::pin_project;
+
+    #[allow(missing_debug_implementations)] // https://github.com/rust-lang/rust/pull/74060
+    #[allow(variant_size_differences)] // for the type itself
+    #[allow(clippy::large_enum_variant)] // for the type itself
```
taiki-e, line 141: Note that it does not suppress lints against the `#[pin_project]` type itself.
However, users can suppress lints simply by allowing it on the type. (Previously it need to be allowed at the module level.)

PR #284, opened 2020-10-01
Allow clippy::redundant_pub_crate

dnaka91:
This is my first PR to this project so please bear with me. 

While using this crate in my project I noticed that the clippy::redundant_pub_crate (part of clippy::nursery) came up when I was using the derive macro in a module within a binary crate.

For example my trimmed down project layout is like this:
```
src/main.rs
src/services/log.rs <-- using the pin_project derive macro here creates a clippy warning
```

This change allows the clippy::redundant_pub_crate lint on all generated code to silence the lint messages. I hope this is the right spot to add the `#[allow(...)]` attribute.


taiki-e (commented):
Thanks! Could you run the following command to fix expandtest failure?
```sh
rm -r tests/expand/tests/expand/*.expanded.rs && cargo +nightly test --manifest-path tests/expand/Cargo.toml
```

dnaka91:
Thank you for the quick reply.
I rebased and updated the expansion tests.

taiki-e (approved):

PR #287, opened 2020-10-04
Fix drop order of pinned fields in project_replace

taiki-e:
The fields of the struct and the active enum variant are dropped in declaration order ([reference](https://doc.rust-lang.org/reference/destructors.html)).

The current `project_replace` drops pinned fields in reverse order of declaration ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5e2117bc64a648f254204d4e858bfb2a)), but this should be in the same order as dropping structs normally.

cc @Diggsey


Aaron1011 (approved):

taiki-e:
bors r=Aaron1011

PR #291, opened 2020-10-05
Fix doctest failure

taiki-e:


taiki-e:
```text
error: unknown lint: `drop_bounds`
 --> src/lib.rs:121:9
  |
6 | #[allow(drop_bounds, unknown_lints)]
  |         ^^^^^^^^^^^
  |
```
...seems that `unknown_lints` needs to be declared first.

PR #292, opened 2020-10-05
Bump MSRV to 1.37

taiki-e:
Bumps MSRV to 1.37 because `underscore_const_names` requires 1.37. This allows removing the build script.

It's been more than a year since [1.37 was released](https://blog.rust-lang.org/2019/08/15/Rust-1.37.0.html), and most of the users of this crate depend on `futures_api` that requires 1.36 and `async_await` that requires 1.39. so bumping MSRV to 1.37 is probably no problem.
Also, given that pin-project-lite's MSRV is 1.37 and the possibility of switching, most futures-related utilities require at least 1.37.

cc #264
Refs: #219


taiki-e:
As far as I know, futures/futures-util/futures-test with no-default-features are the only futures-related utilities that have been tested to compatibility with 1.36 and depend on pin-project.
However, this should be fine as that is undocumented and there is a possibility of switching to pin-project-lite. (also probably many users have the default or async-await feature enabled.)

taiki-e:
bors r+

See also https://github.com/taiki-e/pin-project/issues/264#issuecomment-703690654

PR #308, opened 2021-01-09
Suppress clippy::ref_option_ref lint in generated code

taiki-e:
https://github.com/taiki-e/pin-project-lite/issues/44


taiki-e:
bors r-

I need to update expandtest

PR #321, opened 2021-03-27
Prepare for removal of safe_packed_borrows lint

taiki-e:
Same as https://github.com/taiki-e/pin-project-lite/pull/55, but, pin-project's safety no longer depends on this, so there is no need to rush release.


taiki-e (commented):
bors r+

PR #322, opened 2021-03-28
v0.4: Fix unused_must_use warning on unused borrows

taiki-e:
This fixes `unused_must_use` warning on unused borrows, which will be added to rustc in the future.

See https://github.com/rust-lang/rust/pull/76894 fore more.

(Note: pin-project 1.0 does not have this problem.)


taiki-e (commented):
bors r+

PR #324, opened 2021-04-13
Update ui tests to nightly-2021-04-13

taiki-e:


tests/ui/pin_project/packed-name-value.stderr
```diff
@@ -1,17 +1,52 @@
...
+note: we would appreciate a bug report: https://github.com/rust-lang/rust/issues/new?labels=C-bug%2C+I-ICE%2C+T-compiler&template=ice.md
+
+note: rustc 1.53.0-nightly (d0695c908 2021-04-12) running on x86_64-apple-darwin
+
+note: compiler flags: -C embed-bitcode=no -C split-debuginfo=unpacked -C debuginfo=2 --crate-type bin
+
+note: some of the compiler flags provided by cargo are hidden
+
+query stack during panic:
+end of query stack
```
taiki-e, line 52: :(

taiki-e: It's a bad idea to include an error message in the cross-platform ui test that includes the name of the toolchain and the target, so we need to avoid this.

taiki-e: Looks like https://github.com/rust-lang/rust/issues/83921

taiki-e: So I don't think we should rely on the behavior of rustc that handles this correctly. That is, all `repr(packed ...)` should be rejected on our part.

PR #326, opened 2021-04-16
Preserve `self` as segment of path

taiki-e:
Based on https://github.com/dtolnay/async-trait/commit/6029cbf375c562ca98fa5748e9d950a8ff93b0e7


taiki-e (commented):
bors r+

PR #332, opened 2021-07-30
Install rust-src component for ui tests

taiki-e:
Some tests generate different diagnostics depending on whether this component is installed.


taiki-e:
This resolves the problem that is what has annoyed me in some projects. https://github.com/taiki-e/pin-project/pull/254#issuecomment-643591543

bors r+

PR #354, opened 2023-05-12
Update to syn-2 (Continuation)

daxpedda:
Continuing where @maurer left off.

Do you need me to squash?

Replaces #352.

### DIFF
--- a/tests/ui/pin_project/add-attr-to-struct.rs
+++ b/tests/ui/pin_project/add-attr-to-struct.rs
@@ -4,7 +4,7 @@ use auxiliary_macro::add_pin_attr;
 use pin_project::pin_project;
 
 #[pin_project]
-#[add_pin_attr(struct)] //~ ERROR duplicate #[pin] attribute
+#[add_pin_attr(struct)] //~ ERROR expected attribute arguments in parentheses
 struct Foo {
     #[pin]
     f: PhantomPinned,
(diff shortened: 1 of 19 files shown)


tests/ui/pin_project/add-attr-to-struct.rs
```diff
@@ -4,7 +4,7 @@ use auxiliary_macro::add_pin_attr;
 use pin_project::pin_project;
 
 #[pin_project]
-#[add_pin_attr(struct)] //~ ERROR duplicate #[pin] attribute
+#[pin] //~ ERROR duplicate #[pin] attribute
```
daxpedda, line 7: Should I update comments like this to something more appropriate?

tests/ui/pin_project/add-attr-to-struct.stderr
```diff
@@ -1,4 +1,4 @@
-error: duplicate #[pin] attribute
+error: expected attribute arguments in parentheses: `pin(...)`
```
daxpedda, line 1: Should I update comments like this to something more appropriate?
https://github.com/taiki-e/pin-project/blob/ce860dd624aeabede645f38745aa9af334e559f0/tests/ui/pin_project/add-attr-to-struct.rs#L7

taiki-e: Yes, please update these comments to match new error messages.

daxpedda: Done.
I tried to keep the style consistent, let me know if you want me to change something.

taiki-e (approved):
Thanks!

taiki-e:
Published in 1.1.0.

PR #356, opened 2024-01-25
Add `allow(clippy::missing_docs_in_private_items)`

daxpedda:
Apparently `clippy::missing_docs_in_private_items` triggers in proc-macros, seeing that `pin-project` already adds a sizable amount of `#[allow(...)]`s I thought a PR might not hurt.

I also submitted an issue at Clippy: https://github.com/rust-lang/rust-clippy/issues/12197.

### DIFF
--- a/pin-project-internal/src/pin_project/derive.rs
+++ b/pin-project-internal/src/pin_project/derive.rs
@@ -136,6 +136,7 @@ fn proj_allowed_lints(cx: &Context<'_>) -> (TokenStream, TokenStream, TokenStrea
         #proj_mut_allowed_lints
         #[allow(dead_code)] // This lint warns unused fields/variants.
         #[allow(clippy::mut_mut)] // This lint warns `&mut &mut <ty>`.
+        #[allow(clippy::missing_docs_in_private_items)]
     };
     let proj_ref_allowed_lints = if cx.project_ref { Some(&global_allowed_lints) } else { None };
     let proj_ref = quote! {
@@ -148,6 +149,7 @@ fn proj_allowed_lints(cx: &Context<'_>) -> (TokenStream, TokenStream, TokenStrea
     let proj_own = quote! {
         #proj_own_allowed_lints
         #[allow(dead_code)] // This lint warns unused fields/variants.
+        #[allow(clippy::missing_docs_in_private_items)]
         #large_enum_variant
     };
     (proj_mut, proj_ref, proj_own)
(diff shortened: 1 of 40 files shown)


taiki-e:
Thanks! Could you add a case that triggers this to [tests/lint.rs](https://github.com/taiki-e/pin-project/blob/main/tests/lint.rs) test?

daxpedda:
The lint apparently doesn't trigger in tests, so I can't add it in `tests/lint.rs`.
I tried adding it in a unit test, but that doesn't work either because the proc-macro doesn't work as it can't find `pin_project` as a dependency.
Also tried to add `pin_project` as a `dev-dependency`, but that still requires me to put it behind a `#[cfg(test)]` guard, which again disabled the lint.

In the meantime I found that empty enum variants can still trigger this lint, so I had to add another `#[allow(...)]`.

Here is some code that can trigger both test cases:
```rust
//! Testing `clippy::missing_docs_in_private_items`.

#![warn(clippy::missing_docs_in_private_items)]

use pin_project::pin_project;

#[pin_project(project_replace)]
pub struct Test1 {
    pub test: u8, // Triggered here.
}

#[pin_project(project = Test2Proj)]
pub enum Test2 {
    VariantA(u8),
    VariantB, // Triggered here.
}
```

pin-project-internal/src/pin_project/derive.rs
```diff
@@ -136,6 +136,7 @@ fn proj_allowed_lints(cx: &Context<'_>) -> (TokenStream, TokenStream, TokenStrea
         #proj_mut_allowed_lints
         #[allow(dead_code)] // This lint warns unused fields/variants.
         #[allow(clippy::mut_mut)] // This lint warns `&mut &mut <ty>`.
+        #[allow(clippy::missing_docs_in_private_items)]
```
daxpedda, line 139: I could add some code that only inserts this in the case of an enum?

We can't fold it into `large_enum_variant` because it's also required in structs in `proj_own`, so that would duplicate the lint.

daxpedda:
I guess one last idea to get this lint tested would be to make a sub-crate only for lint testing, so it's not behind a `#[cfg(test)]`.
Could potentially also move the whole of `tests/lint.rs` there?

taiki-e:
> I guess one last idea to get this lint tested would be to make a sub-crate only for lint testing, so it's not behind a `#[cfg(test)]`.
> Could potentially also move the whole of `tests/lint.rs` there?

I think this (convert the `tests/lint.rs` to a new test crate) is a reasonable idea. We already have such test crates to check compatibility with no-std/old edition.

- https://github.com/taiki-e/pin-project/tree/main/tests/no-std
- https://github.com/taiki-e/pin-project/tree/main/tests/rust-2015

daxpedda:
Done, this also uncovered two cases of FP that I addressed as well.

taiki-e (approved):
Thanks!

taiki-e:
Published in 1.1.4.

PR #357, opened 2024-10-24
Work around negative_impls that allows unsound overlapping Unpin impls

### LINKED ISSUE #340: #![feature(negative_impls)] breaks the guarantee for Unpin provided by pin-project
pin-project provides an appropriate Unpin implementation by default. Since overlapping implementations are prohibited, this ensures that users cannot add inappropriate Unpin implementations.

However, currently, this guarantee can be broken by using [`#[feature(negative_impls)]`](https://github.com/rust-lang/rust/issues/68318): [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=d2e87f361ddd443608401b5d10a2b450)

Thanks @danielhenrymantilla  for [pointing out the interaction with this feature](https://rust-lang.zulipchat.com/#narrow/stream/213817-t-lang/topic/design.20meeting.3A.20backlog.20bonanza/near/269471299).

taiki-e:
See https://github.com/taiki-e/pin-project/issues/340#issuecomment-2432146009 for details of the workaround.

(Btw, lifetime to workaround trivial_bounds is still needed.)

Fixes #340

### DIFF
--- a/examples/enum-default-expanded.rs
+++ b/examples/enum-default-expanded.rs
@@ -77,13 +77,15 @@ const _: () = {
         __field0: T,
     }
     impl<'pin, T, U> ::pin_project::__private::Unpin for Enum<T, U> where
-        __Enum<'pin, T, U>: ::pin_project::__private::Unpin
+        ::pin_project::__private::PinnedFieldsOf<__Enum<'pin, T, U>>:
+            ::pin_project::__private::Unpin
     {
     }
     // A dummy impl of `UnsafeUnpin`, to ensure that the user cannot implement it.
     #[doc(hidden)]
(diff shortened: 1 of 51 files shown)


danielhenrymantilla:
> (Btw, lifetime to workaround trivial_bounds is still needed.)

Yeah, the other approach, but not worth changing your code for at the moment, would have been to emit `for<'__trivial> Type<…> : Bounds…`. That avoids the "trivially false" errors, much like I did over #340 with my `for<'never> str : Sized`.

This is more of an FYI should the current "unconstrained lifetime" be to become an error in some future version of Rust, again, not suggesting that you change the current "floating" `'__pin` design

PR #362, opened 2025-08-24
Add pub prefix to args to make the method and the returned struct

### LINKED ISSUE #361: Allow pub visibility as an option
I would like, eventually as an option, to make the .project method and ProjectionRef struct public

My use case is a user-defined callback, tacking as argument a Pin<&mut MyStruct>

MyStruct have some structurally pinned fields that the user probably do not want to touch, but also regular fields designed to be modified by the user, and I do not want to pass them directly to the user (aka, projecting on the library side) to allow the user to call the method defined on Pin<&mut MyStruct> (which need some of the structurally pinned fields)

It may be good idea to require the use of the project = SomeName option in that case, and maybe a way to add doc to the ProjectRef and the project method (or a good default doc)

However, as they are both tagged doc(hidden), it may be just as easy to just make them follow the visibility of the item they are defined on (in case there is another reason besside not appearing in the doc i am glad to hear it)

I would be happy to make a PR for whatever behavior you think is best! (in the meantime, I think I will use a personal fork where I implement the easier "follow whatever the item visibility is" option)

louwenus:
Behavior on existing syntax is kept as is.
The new syntax look like this:

```rust
# use pin_project::pin_project
#[pin_project(pub project = StructProj)] //force the project method and StructProj to be pub
pub struct Struct<T> {
    #[pin]
    field: T,
}
```
pub prefix can also be used on project_ref and (named) project_replace

fix #361

### DIFF
--- a/pin-project-internal/src/pin_project/args.rs
+++ b/pin-project-internal/src/pin_project/args.rs
@@ -3,9 +3,7 @@
 use proc_macro2::{Span, TokenStream};
 use quote::quote;
 use syn::{
-    Attribute, Error, Ident, Result, Token,
-    parse::{Parse, ParseStream},
-    spanned::Spanned as _,
+    parse::{Parse, ParseStream}, spanned::Spanned as _, Attribute, Error, Ident, Result, Token, Visibility
 };
 
 use super::PIN;
@@ -64,11 +62,11 @@ pub(super) struct Args {
     /// `UnsafeUnpin` or `!Unpin` argument.
     pub(super) unpin_impl: UnpinImpl,
     /// `project = <ident>` argument.
-    pub(super) project: Option<Ident>,
+    pub(super) project: ProjArgs,
     /// `project_ref = <ident>` argument.
-    pub(super) project_ref: Option<Ident>,
+    pub(super) project_ref: ProjArgs,
     /// `project_replace [= <ident>]` argument.
-    pub(super) project_replace: ProjReplace,
+    pub(super) project_replace: ProjArgs,
 }
 
 impl Parse for Args {
@@ -103,11 +101,19 @@ impl Parse for Args {
         let mut unsafe_unpin = None;
         let mut not_unpin = None;
         let mut project = None;
+        let mut project_pub = false;
         let mut project_ref = None;
+        let mut project_ref_pub = false;
         let mut project_replace_value = None;
         let mut project_replace_span = None;
+        let mut project_replace_pub = false;
+        let mut visibility_pub = None;
 
         while !input.is_empty() {
+            if input.peek(Token![pub]) {
+                let pub_token: Token![pub] = input.parse()?;
+                visibility_pub.replace(pub_token.span());
+            }
             if input.peek(Token![!]) {
                 let bang: Token![!] = input.parse()?;
                 if input.is_empty() {
@@ -133,9 +139,11 @@ impl Parse for Args {
                     }
                     "project" => {
                         project = Some(parse_value(input, &token, project.is_some())?.0);
+                        project_pub = visibility_pub.take().is_some();
                     }
                     "project_ref" => {
                         project_ref = Some(parse_value(input, &token, project_ref.is_some())?.0);
+                        project_ref_pub = visibility_pub.take().is_some();
                     }
                     "project_replace" => {
                         if input.peek(Token![=]) {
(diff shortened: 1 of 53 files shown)


tests/ui/pin_project/invalid.rs
```diff
@@ -193,6 +193,18 @@ mod pin_project_argument {
...
+    struct Pub1(#[pin] ());
+
+    #[pin_project(pub Unpin)] //~ Error `pub` can only be used on project, project_ref or named project_replace.
+    struct Pub2(#[pin] ());
+
+    #[pin_project(pub project_replace)] //~ Error project_replace cannot be pub if it is not named
+    struct Pub3(#[pin] ());
+
+    #[pin_project(pub project_replace = Pub4ProjReplace, pub Project = Pub4Proj)] // Ok
+    struct Pub4(#[pin] ());
```
taiki-e, line 207: The errors here seem all mismatch with ones specified by comments.

```
error: unexpected end of input, expected identifier
   --> tests/ui/pin_project/invalid.rs:197:5
    |
197 |     #[pin_project(pub)] //~ Error `pub` can only be used on project, project_ref or named project_replace.
    |     ^^^^^^^^^^^^^^^^^^^
    |
    = note: this error originates in the derive macro `::pin_project::__private::__PinProjectInternalDerive` (in Nightly builds, run with -Z macro-backtrace for more info)

error: unexpected argument: Unpin
   --> tests/ui/pin_project/invalid.rs:200:23
    |
200 |     #[pin_project(pub Unpin)] //~ Error `pub` can only be used on project, project_ref or named project_replace.
    |                       ^^^^^

error: `pub` can only be used on project, project_ref or named project_replace.
   --> tests/ui/pin_project/invalid.rs:203:19
    |
203 |     #[pin_project(pub project_replace)] //~ Error project_replace cannot be pub if it is not named
    |                   ^^^

error: unexpected argument: Project
   --> tests/ui/pin_project/invalid.rs:206:62
    |
206 |     #[pin_project(pub project_replace = Pub4ProjReplace, pub Project = Pub4Proj)] // Ok
    |                                                              ^^^^^^^
```

pin-project-internal/src/pin_project/args.rs
```diff
@@ -190,10 +205,28 @@ impl Parse for Args {
...
+            match (project_replace_span, project_replace_value, project_replace_pub) {
+                (None, _, _) => ProjArgs::None,
+                (Some(span), Some(ident), false) => ProjArgs::Named { ident, span },
+                (Some(span), Some(ident), true) => ProjArgs::NamedPublic { span, ident },
+                (Some(span), None, false) => ProjArgs::Unnamed { span },
+                (Some(span), None, true) => {
+                    return Err(Error::new(
+                        span,
+                        "project_replace cannot be pub if it is not named",
+                    ));
```
taiki-e, line 218: If we reach this point, it's a bug in our code, not the user's fault, so `unreachable!` would be better here.

taiki-e (changes requested):
Thanks! This looks good to me.

Could you also add `tests/expand` test case for `pub project_replace = ...`?

taiki-e (changes requested):
Thanks.

I noticed you also mentioned about documentation in #361. For now, I think it's okay to use `allow(missing_docs)` (or simple docs), but could you add test cases to `tests/include/basic-safe-part.rs` to see if it works?

louwenus:
I added some simple doc for the structs and methods (feel free to correct/modify them(you should have edit rigths) and/or tell me what to change if they are unclear/incomplete/have grammar errors ...)

for the fields, I copy the referenced field doc without any change.

also, the `#[allow]` where not copied over from the struct definition to the project struct, leading to clippy errors (about public not non-exhaustive structs)

For now i simply added a #[non-exhaustive] to the generated struct / enum, which seem a bit hacky to me but quite reasonable as we may want to add some fields of our own (like some PhantomData maybe), and solve the immediate problem of passing the CI

As long as the user is reasonable, as in, do not use some name that need `#[allow]`, i do not see many other use-case where the public struct could need some attribute to make it correct for the public interface, but i could be wrong

A more clean solution could be to copy some attribute and/or add a syntax to add attributes to the generated structs (which could be used for other purposes like derive(Debug) and similar ). What do you think about it ? (I could do that in a following PR if you want)

